PR #26, opened 2016-04-11
Remove invalid function parameters.

varunpatro:
Because of GPU.js's strict restrictions, the function passed to the kernel cannot have variable number of arguments passed to it.

Hence, either the `render` function should be called with a value or the function passed to the `createKernel` method should not accept any arguments. Because the function does not require any arguments, it's signature should not require any arguments as well.


fuzzie360:
Thanks for taking the time to fix the typo.

PR #56, opened 2016-12-25
fix(isWebgl): return a boolean

cazala:
This is to make `gpu.supportWebgl()` return a boolean instead of a function.

gpu

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -355,7 +355,7 @@ var GPUUtils = (function() {
 	function isWebgl( webglObj ) {
 		return (
 			webglObj != null &&
-			webglObj.getExtension
+			!!webglObj.getExtension
 		);
 	}
 	GPUUtils.isWebgl = isWebgl;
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -355,7 +355,7 @@ var GPUUtils = (function() {
 	function isWebgl( webglObj ) {
 		return (
 			webglObj != null &&
-			webglObj.getExtension
+			!!webglObj.getExtension
```
robertleeplummerjr: Why not just: `webglObj.hasOwnProperty('getExtension')`

robertleeplummerjr: More characters, but no truthy, only truth.

robertleeplummerjr:
@cazala is on fire!

PicoCreator:
@cazala & @robertleeplummerjr - Thanks for the pull, sorry for the delay. It has been merged in. 

Cheers =)

robertleeplummerjr:
(it was a pleasure)

PR #65, opened 2017-04-20
Fix flatten performance for larger array sizes

### LINKED ISSUE #64: Errors with very large arrays
Hi,

When trying to execute a GPU function on a large array (1920x1080x3 which represents a 1920x1080 picture with 3 color components) it fails with the following error:
"RangeError: too many arguments provided for a function call" on Firefox
"RangeError: Maximum call stack size exceeded" on Chrome.

See attached example, which do has non-op function to exhibit the issue: https://drive.google.com/file/d/0BxAjgoe3PaRwaTFHMUsxWG1xV2c/view?usp=sharing
It works when working on smaller arrays, but fails on large ones.

I'm not sure how big arrays are common in Javascript (I feel like Firefox or Chrome have issues handling them anyway) but such arrays don't sound that big to me, particularly for taking full advantage of GPU operations by working on big data set.

robertleeplummerjr:
Fixes #64.

### DIFF
--- a/src/backend/mode_gpu.js
+++ b/src/backend/mode_gpu.js
@@ -57,24 +57,34 @@
 		return ret;
 	}
 
-	function flatten(arr) {
-		if (GPUUtils.isArray(arr[0])) {
-			if (GPUUtils.isArray(arr[0][0])) {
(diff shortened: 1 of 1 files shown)


robertleeplummerjr:
housekeeping

PR #66, opened 2017-05-08
Experimental node es6

robertleeplummerjr:
This also adds cpu kernel writer, which significantly improves speed on cpu.  In-fact, side by side it'll be many times faster than native javascript.


robertleeplummerjr:
This is currently failing for cpu, but I'm curious if it should be supported:
```js
var gpu = new GPU({ mode: mode });
	
	var f = gpu.createKernel(function(a, b) {
		function custom_adder(a,b) {
			return a+b;
		}
		
		return custom_adder(a[this.thread.x], b[this.thread.x]);
	}, {
		dimensions : [6]
	});
```

If I were to pseudoscript this, it'd look like:
"for each first make a function and then use it to add a to b in that function"
```
new Array(1000000).forEach((a) => {
  function add(a, b) { return a + b; }
  return add(a, 100);
});
```

It just seems confusing.

robertleeplummerjr:
As a follow up, the already implemented and seemingly more refined (imho):
```js
var gpu = new GPU({ mode: mode });
gpu.addFunction(function custom_adder(a,b) {
  return a+b;
});
var f = gpu.createKernel(function(a, b) {
  return custom_adder(a[this.thread.x], b[this.thread.x]);
}, {
  dimensions : [6]
});
```

is not only easier to understand, call, or look at, it already works perfectly.

robertleeplummerjr:
The next item is: what instance of what object owns the canvas and webgl context?  Should it be the instance of the kernel, the runner, or of GPU itself?

robertleeplummerjr:
All unit tests with the exception of the above comments are running correctly, fyi.

robertleeplummerjr:
note: still need to get graphical working

robertleeplummerjr:
note: graphical working

robertleeplummerjr:
With insight from https://gitter.im/gpujs/gpu.js?at=59171504631b8e4e61d09d05, closing, and will reopen on `develop` branch.

PR #104, opened 2017-06-30
CPU fallback and tests fix

abhisheksoni27:


src/gpu.js
```diff
@@ -130,6 +130,13 @@ module.exports = class GPU {
 	/// 	{Function} callable kernel function
 	///
 	createKernels() {
+		if (!utils.isWebGlDrawBuffersSupported) {
```
robertleeplummerjr: This needs to move a few more lines down, and needs to accept settings.

abhisheksoni27: Moved. 

Also, The `utils.isWebGlDrawBuffersSupported` method, as it is defined in `utils`, doesn't take in any `settings`? 

It's a getter for a field. [utils.js | Line358](https://github.com/gpujs/gpu.js/blob/develop/src/utils.js#L358)

What did you mean by "needs to accept settings"? :thinking:

robertleeplummerjr:
very nice sir!

Issue #110, opened 2017-07-03
Documentation theme selection


abhisheksoni27:
This is how [docdash](https://github.com/clenemt/docdash) looks:

![gpu js](https://user-images.githubusercontent.com/4340071/27803974-f69c2398-6049-11e7-97d6-401922ac1326.png)

And this is [minami](https://github.com/Nijikokun/minami):

![gpu js-minami](https://user-images.githubusercontent.com/4340071/27803977-ffa1b520-6049-11e7-8e54-5b3f1f9c3abb.png)

I think **docdash** is cool. We can change the theme color, and the nested sidebar output is great.

@robertleeplummerjr and @PicoCreator What do you think?

robertleeplummerjr:
I believe docdash has my vote.

robertleeplummerjr:
Maybe in the "official" color scheme here: ![](https://camo.githubusercontent.com/f48abdd17dfd3a682a55b5f3a9be83bba06e5c7d/687474703a2f2f6770752e726f636b732f696d672f6f67696d6167652e706e67)

abhisheksoni27:
@robertleeplummerjr Good Point, Sir. :tada: 

How can we inject custom `CSS` to auto-generated docs? 

`docdash` doesn't currently have an option to adjust colors using options.

abhisheksoni27:
[Ability to add custom styles](https://github.com/clenemt/docdash/issues/19#issuecomment-254195869)

That's a hack! But it works for us. We just need to fork it to `gpujs`, and change the colors.

PicoCreator:
I think CSS / JS injection can pretty much let us get away with anything.... even crazy changes haha

As long as the core data formatting is valid, and there

abhisheksoni27:
We can write a gulp task to write a new **CSS** file in the generated docs folder, right after `jsdoc` finishes. And in the `build.xml` file, we can add that gulp task. 

I think that would work. What do you think?

abhisheksoni27:
This is complete.

PR #111, opened 2017-07-04
JSDoc is here

abhisheksoni27:
@robertleeplummerjr This is using [docdash](https://github.com/clenemt/docdash). As you pointed out in #110, we should change the theme color.

For that, we can either inject `CSS/JS`, or we can maintain our own copy of `docdash` and pull that everytime `setup-docs` is called.

We could fork the project to `gpujs` (parent organization) and I could change the color themes in that copy.(Though, that's my opinion.)

If you have any other idea, please share.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -20,32 +20,14 @@
 	</target>
 	
 	<target name="build-docs">
-		<mkdir dir="doc/dev-docs"/>
-		<mkdir dir="doc/usr-docs"/>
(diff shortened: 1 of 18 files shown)


PicoCreator:
I think CSS / JS injection can pretty much let us get away with anything.... even crazy changes haha

As long as the core data formatting is valid, and there

PR #112, opened 2017-07-05
CSS injection using gulp to modify docdash's theme

abhisheksoni27:
#110

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -29,6 +29,10 @@
 			<arg value="doc"/>
 			<arg value="--debug"/>
 		</exec>
+
+		<exec executable="gulp">
+			<arg value="injectCSS"/>
+		</exec>
 	</target>
 	
 	<!-- **************************************************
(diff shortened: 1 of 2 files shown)


PicoCreator:
Minor one, Fixed the missing NPM modules that were required - see : https://github.com/gpujs/gpu.js/pull/124

Additionally, if it makes it easier, you can abandon the ant build.xml, the main reason why it was there is that it was easier to setup a git-clone hack for natural docs. Which you are deprecating. 

So feel free to do the whole build process via npm / grunt =)

abhisheksoni27:
Great. :tada: 

I was wondering why we were using `gulp and ant` for the docs and setup when we could do it just using `gulp`. Thanks for clarifying that. I'll migrate the build process to `gulp` only. (With help from `npm scripts`)

Also, as it stands, the dependencies you added were part of the build process. I mean, they weren't included by default because not of all of the contributors would be working with docs. But it's okay. I will see what it's like migrating away from ant, and will update the dependencies accordingly. :smile:

PR #139, opened 2017-07-15
Fix kernel to setup parameters again and refactor isArray

abhisheksoni27:


src/core/utils.js
```diff
@@ -243,21 +243,11 @@ class Utils extends UtilsCore {
...
-			case Uint16Array:
-			case Int32Array:
-			case Uint32Array:
-			case Float32Array:
-			case Float64Array:
-			case Array:
-				return true;
-			default:
-				return false;
+		if (array.length) {
```
robertleeplummerjr: I don't think this works on two accounts:
`utils.isArray('hi mom') -> true`
`utils.isArray([]) -> false`

abhisheksoni27: Should we worry about the first one? :laughing: 

Fixed.

src/core/utils.js
```diff
@@ -243,21 +243,11 @@ class Utils extends UtilsCore {
...
-			case Uint16Array:
-			case Int32Array:
-			case Uint32Array:
-			case Float32Array:
-			case Float64Array:
-			case Array:
-				return true;
-			default:
-				return false;
+		if((array.length || array.length === 0) && (array.constructor !== String)) {
```
robertleeplummerjr: How about: `if (array.hasOwnProperty('length'))`

abhisheksoni27: Good catch. But it, unfortunately, fails with `Strings`.

robertleeplummerjr: Really?  Typed arrays?

robertleeplummerjr: wow, mind blown

robertleeplummerjr: What do you think about keeping this existing functionality, and adding a method that allows them to add their array type, that could iterate through, and if it exists we could return true.  That way it is all strongly typed.

abhisheksoni27: Indeed. 
screenshot from 2017-07-15 23-45-03

robertleeplummerjr: I ran it locally, and my mind exploded.

abhisheksoni27: I am not sure I totally get what you are trying to say here. Can you elaborate the part where you say "that allows them to add their array type"?

abhisheksoni27: I know, right? When I saw your initial comment, I was like, cool that's neat and then I ran it. :laughing:

robertleeplummerjr: Like this:
```js
static isArray(array) {
		switch (array.constructor) {
			case Int8Array:
			case Uint8Array:
			case Uint8ClampedArray:
			case Int16Array:
			case Uint16Array:
			case Int32Array:
			case Uint32Array:
			case Float32Array:
			case Float64Array:
			case Array:
				return true;
			default:
for (let i = 0; i < arrayTypes.length; i++) {
  if (array.constructor === arrayTypes[i]) {
    return true;
  }
}
				return false;
		}
}
```

abhisheksoni27: And we can store the different types in an Array itself, rather than using a `switch` statement. Got it,

robertleeplummerjr: You think it is worth it?

abhisheksoni27: It's not. It goes over the array every time, which is inefficient.
If we use `.find`, it does the same thing behind the scenes.

robertleeplummerjr: Got it!
```js
static isArray(array) {
		switch (array.constructor) {
			case Int8Array:
			case Uint8Array:
			case Uint8ClampedArray:
			case Int16Array:
			case Uint16Array:
			case Int32Array:
			case Uint32Array:
			case Float32Array:
			case Float64Array:
			case Array:
				return true;
			default:
				return Array.isArray(array);
		}
}
```

Booya!

PR #176, opened 2017-08-31
175-fix

Drulac:


src/backend/runner-base.js
```diff
@@ -94,16 +94,23 @@ module.exports = class BaseRunner {
...
 	 * @desc Get and returns the Synchronous executor, of a class and kernel
 	 * Which returns the result directly after passing the arguments.
-	 * 
+	 *
 	 */
 	buildKernel(fn, settings) {
 		settings = Object.assign({}, settings || {});
-		const fnString = fn.toString();
+
+		var fnString = '';
```
robertleeplummerjr: `const fnString = fn.toString();` should be sufficient on a string or a function.

src/core/gpu.js
```diff
@@ -83,7 +83,7 @@ class GPU extends GPUCore {
 		if (typeof fn === 'undefined') {
 			throw 'Missing fn parameter';
 		}
-		if (!utils.isFunction(fn)) {
+		if (!utils.isFunction(fn) && typeof fn !== 'string') {
```
robertleeplummerjr: The logic that checks for `isFunction` should probably contain the logic for string or not.  Can you more that there?

src/backend/runner-base.js
Drulac: I will do that

src/core/gpu.js
Drulac: No, if fn is a string, that will fire the if and throw an error

PR #179, opened 2017-09-03
Fix #177

### LINKED ISSUE #177: I got error 'gpu.min.js:16 Uncaught Auto dimensions only supported for kernels with only one input'
I got an error when multiplying matrices.

matrix shape are [4194304, 1] and [1, 4194304].

And I wrote code as following... 

```js
function matMult(a, b, x_dim, y_dim) {
    console.log(typeof(a), typeof(b), a, b)
    const opt = {
        output: {
            x: x_dim,
            y: y_dim
        }
    };

    func = gpu.createKernel(function() {
        var sum = 0;
        for (var i = 0; i < x_dim; i++) {
            sum += a[this.thread.y][i] * b[i][this.thread.y];
        }
        return sum;
    }, opt);
    return func(a, b)
}

var r = math.sqrt(matMult(x.valueOf(), math.transpose(y).valueOf(), 1, n_point))
```

sorry, Im begginer, so I cannot solve myself.

RationalCoding:
Flattens the `output` object when it is passed as an option.


robertleeplummerjr:
Awesome!  Can you run `gulp beautify && gulp build`?

RationalCoding:
Done!

robertleeplummerjr:
Nice job!

PR #189, opened 2017-09-20
Add jsDelivr link

LukasDrgon:
[jsDelivr](https://www.jsdelivr.com/) is the [fastest opensource cdn](https://www.cdnperf.com/) available and built specifically for production usage. It can serve any project from npm with zero config. This PR adds it to readme and it also changes the default file to `bin/gpu.min.js`. These changes will only work after you delete bin/*.js from `.npmignore`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,10 +67,10 @@ yarn add gpu.js
 
 ### Browser
 
-Download the latest version of gpu.js and include the files in your HTML page using the following tags:
+Include the files in your HTML page using the following tags:
 
 ```html
-<script src="/path/to/js/gpu.min.js"></script>
+<script src="//cdn.jsdelivr.net/npm/gpu.min.js/bin/gpu.min.js"></script>
 ```
 
 In JavaScript, initialize the library:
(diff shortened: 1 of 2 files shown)


robertleeplummerjr:
@fuzzie360 or @PicoCreator, any feedback here?

robertleeplummerjr:
Sorry for the weight, the celebrities are very busy.

fuzzie360:
Thanks @LukasDrgon for informing us about jsDelivr. Seems like a good idea to adopt this CDN, but I would prefer to word the documentation in such a way that you have a choice between self-host and CDN.

Also, I believe we must get versioning down before we make full use of this CDN. So I suppose we should just bite the bullet and release v1.0 and start logging our releases soon.

LukasDrgon:
In that case i suggest you publish 1.0.0 first and i will update this PR afterwards.

robertleeplummerjr:
@LukasDrgon ty for your diligence here.

Fil:
would be cool to also have it under `https://unpkg.com/gpu.js` (without any more complexities). The link to add to package.json would simply be
`"unpkg": "bin/gpu.min.js",`

PR #192, opened 2017-09-29
191 js compiler

robertleeplummerjr:
js in... js out


robertleeplummerjr:
Going to merge in because I'm impatient.  But will gladly roll back if there are objections.

PR #206, opened 2017-10-17
This allows for flat arrays to be read in as multi dimensional arrays

robertleeplummerjr:


robertleeplummerjr:
I minor improvement in performance.

image

robertleeplummerjr:
Something I've been tossing around in my head is how better to send the variable in, something that comes to mind is perhaps using the argument name as a possible indicator, rather than calling a function on the argument directly.

Example:

```js
const gpu = new GPU();
const kernel = gpu.createKernel(function(matrix1d) {
});
```

Would essentially be used to call:
```js
const gpu = new GPU();
const kernel = gpu.createKernel(function(GPU.input(matrix)) {
});
```

Too, we still need compatibility on the CPU side.  What this is telling GPU.js:  `matrix1d` is a flat array, that may have 1,2 or 3 dimensions, but it isn't converted to a flattened array.  I'm open to other naming conventions as well.

robertleeplummerjr:
if we do add support for syntax of variable, it'll be in a later pr

PR #221, opened 2017-11-08
Modified README to include list of supported functions in the JavaScript Math module

dannycalleri:
Just some slight editing done on the README. Added an anchor in the summary and the corresponding section.


robertleeplummerjr:
great work!

Issue #58, opened 2017-02-25
Safari support?


brianxautumn:
When I try to run on safari I get:
WebGL: INVALID_ENUM: readPixels: invalid type

fuzzie360:
Thanks for the report.

The issue is Safari does not support readPixels with float type. There is already code to detect if the browser supports this, but it seems I have been broken recently. I will fix it soon.

robertleeplummerjr:
Safari is supported, with the limitation of kernel maps.

robertleeplummerjr:
via: https://github.com/gpujs/gpu.js/pull/234#issuecomment-352312020
>I don't think it should be fixed in this manner.

Reopening to try and find a quick fix for v1

robertleeplummerjr:
Let's get this resolved.  Here is a quick and dirty possibility to skirt this issue https://github.com/gpujs/gpu.js/pull/235

robertleeplummerjr:
The texture issue is resolved as of https://github.com/gpujs/gpu.js/releases/tag/1.0.0, however there is some sort of texture size issue that needs to be resolved before this can be closed.

robertleeplummerjr:
Now supported, with limits, in v2 rc13.

robertleeplummerjr:
Even unit tests pass, yay!

PR #234, opened 2017-12-15
Fix floatOutput in createKernelMap

### LINKED ISSUE #233: floatOutput doesn't work with createKernelMap
Example:
```js
var km = new GPU().createKernelMap([
    function add(v1, v2) {
      return v1 + v2;
    },
    function divide(v1, v2) {
      return v1 / v2;
    }
  ], function (a, b, c) {
    return divide(add(a[this.thread.x], b[this.thread.x]), c[this.thread.x]);
}, {output: [3], floatOutput: true})
```

It produce the next glsl:
```glsl
...
void main(void) {
  index = floor(vTexCoord.s * float(uTexSize.x)) + floor(vTexCoord.t * float(uTexSize.y)) * uTexSize.x;
  index *= 4.0;
  threadId = indexTo3D(index, uOutputDim);
  kernel();
  gl_FragColor.r = kernelResult;
  index += 1.0;
  threadId = indexTo3D(index, uOutputDim);
  kernel();
  gl_FragColor.g = kernelResult;
  index += 1.0;
  threadId = indexTo3D(index, uOutputDim);
  kernel();
  gl_FragColor.b = kernelResult;
  index += 1.0;
  threadId = indexTo3D(index, uOutputDim);
  kernel();
  gl_FragColor.a = kernelResult;
}
...
```

That's obviously wrong

loyd:
Fixes #233


loyd:
I have some questions:
1. Why aren't `floatOutput` and `floatTextures` enabled by default?
2. Why is `floatTextures` the option of kernel, not `Texture`?

robertleeplummerjr:
I believe they were added after the initial hackathon, and this just had not been addressed yet.  Perhaps now is the time.  Could you add some detail if we change the questions to:

* Why should floatOutput and floatTextures be enabled by default?
* Why should floatTextures be an option of the texture, and not the kernel?

fuzzie360:
I believe floatOutput and floatTextures should actually be enabled automatically when available for maximum performance and precision.

@robertleeplummerjr 

This line change written for the fix for #58 

https://github.com/gpujs/gpu.js/blob/b6163c1090f5441beb58c29cf371b25643bd13fb/src/backend/web-gl/kernel.js#L78

I don't think it should be fixed in this manner.

https://github.com/gpujs/gpu.js/blob/b6163c1090f5441beb58c29cf371b25643bd13fb/src/backend/web-gl/kernel.js#L70

The purpose of this line is to detect if readPixels support float textures (in Safari)

robertleeplummerjr:
I'll go ahead and merge this in, and since we are so close to the end of the year, and I really want to get v1 out before then, I'll put this as first priority.  Ty guys for you all your hard work!

loyd:
Oh yeah, I forgot about the questions:

> Why should floatOutput and floatTextures be enabled by default?

I agree here w/ @fuzzie360 

> Why should floatTextures be an option of the texture, and not the kernel?

The method of reading or writing (encode to unsigned or not) depends on type of the texture. You cannot write float values to an unsigned texture and vica versa. The current schema has disadvantages:
— You should pass everywhere `floatTextures: true` or `floatTextures: false`, no mixing.
— Also, you can't mix float and unsigned textures (frankly, I don't know why this is necessary).
— You can't use `Texture#toArray()` on a float texture.

In general, textures should contain the type information (in addition to 'float' so desired `vec2`, `vec3`, `ivec2` etc).

PR #235, opened 2017-12-18
propose changes for fix how floating values are handled

robertleeplummerjr:


fuzzie360:
Hi @robertleeplummerjr.

https://github.com/gpujs/gpu.js/blob/2cb67c4244e075be92e3f8d6aa9724df72436a44/bin/gpu.js#L1525-L1526

I think these settings will cause these lines to throw on runtime by default. Which actually now that I've read it, i think I see some logic errors in the code.

https://github.com/gpujs/gpu.js/blob/2cb67c4244e075be92e3f8d6aa9724df72436a44/bin/gpu.js#L2814-L2823

The original behaviour is that there are always 3 states for any setting: on, off and auto. The user can force it to be on or off, or the user can just leave it undefined, and the runtime will pick the best settings always. Which is why I believe they should stay as null/undefined so that they can be defaulted at runtime.

isFloatReadPixelsSupported is intended to tests if readPixels supports float textures, which as you can see is the case for Safari browser even though you can use floating point textures as textures in the bindTexture API. It is not covered by the `OES_texture_float` extension spec so this must be checked by using a kernel launch.

For the test to work properly we need to ensure that the test kernel we launch *do not recursively call the test kernel again*. Actually this is already violated in this version because utils.isFloatReadPixelsSupported() will eventually call utils.isFloatReadPixelsSupported() again from trying to launch the kernel

This is the original version for reference:

https://github.com/gpujs/gpu.js/blob/ba9a51386d6f0c7b5d68495297ae2ef8b6dfe64f/src/backend/mode_gpu.js#L143-L150

I'm sorry for the lack of documentation and readability issues in the if branches, but as you can see the test is not executed by default, and only executed in specific cases. Namely when floatOutputForce is off (previously there was no floatOutput option exposed).

fuzzie360:
Also here is the intended logic for the settings.

floatTextures | floatOutput | OES_texture_float available | isFloatReadPixelsSupported | features enabled
--- | --- | --- | --- | ---
not set | not set | yes | yes  | textures, output
not set | not set | yes | no  | textures
not set | not set | no | any  | -
true | not set | yes | yes  | textures, output
true | not set | yes | no  | textures
true | any | no | any |  throw
true | false | yes | no  | textures
true | true | yes | yes  | textures, output
true | true | yes | no  | throw
false | true | any | any  | invalid config??? maybe throw?

I feel like it was much simpler when floatOutput was not exposed, but I guess its more complete this way.

Also Texture object now needs to be aware if it is a float texture or not.

robertleeplummerjr:
Fantastic feedback, I'll revert and ensure this gets out.  Part of the issue was my own nieve understanding of this library, but this really helps.

robertleeplummerjr:
How does this latest change look?

fuzzie360:
@robertleeplummerjr thanks for the quick work! 

Please see in isFloatReadPixelsSupported:

https://github.com/gpujs/gpu.js/blob/1b74546ea53e329a0dac4062b2376199a0f8d34f/src/core/utils.js#L312-L321

I'm not sure about this, but it seems like isFloatReadPixelsSupported calls createKernel and run which will call build and call validateOptions which will call isFloatReadPixelsSupported. Can you help me check if this is the case?

robertleeplummerjr:
This isn't the case, the logic handles that.

Here is just after we do the check, we set `_isFloatReadPixelsSupported` once, and because it is private causes this not to be called again.
screen shot 2017-12-22 at 10 51 38 am


Here is Chrome's passing unit test:
screen shot 2017-12-22 at 10 48 50 am


Here is Firefox's passing unit test:
screen shot 2017-12-22 at 10 53 36 am


And of course, Safari's nearly passing unit test (better than before):
screen shot 2017-12-22 at 10 54 34 am


The failing Safari unit test has to do with texture size I believe.  It is unrelated, but I will probably need help resolving it on another issue.

fuzzie360:
My bad @robertleeplummerjr ! I see the WebGLValidatorKernel has a different validateOptions.

LGTM on all the changes on #235 and #236. (and merry christmas!)

PR #286, opened 2018-04-27
Support for the do-while construct

galatolofederico:
The implementation is basically the same as the while loop but the js is translated in something like this (pseudocode):
`
for(i = 0;i < LOOP_MAX;i++){
    [BODY]
    if(not [TEST]){
        break;
   }
}
`


robertleeplummerjr:
Fantastic!  Ty for the the addition.

Issue #251, opened 2018-01-30
[Bug?] this.thread.z can be off by 1 with small 3d arrays


saadtazi:
While writing some tests, I found the following:

```const gpu = new GPU();

const k = gpu.createKernel(function () {
  return this.thread.z;
}).setOutput([5, 25, 4]);

console.log(k());
```
Notice that the first number in the 2nd row is `0` instead of `1`.

This seems to happen with small array, and making the `x` dimension a even number "fixes" the issue.

I am pretty new to GPU calculation, so I probably missed some important rule ("dimensions should match this criteria..."), so feel free to close if it is expected behaviour.

http://jsbin.com/diwemoq/edit

Note: I am using gpu.js to run calculation on small arrays... I found this issue while writing some simple tests...

robertleeplummerjr:
@fuzzie360 we are going to need your help here bud

saadtazi:
Looks like it is related to https://github.com/gpujs/gpu.js/issues/241

robertleeplummerjr:
I believe this may have been resolved in https://github.com/gpujs/gpu.js/releases/tag/1.4.8

Can you check the latest?

robertleeplummerjr:
Closing, will reopen if issue persists.

Issue #241, opened 2017-12-30
CPU vs. GPU test maps output differently.


BryanGrezeszak:
I came across this issue while basically just familiarizing myself with the library (checking out how inputs get mapped to outputs via `this.thread.x` and `this.thread.y` and such..._yah, I know it's December 29th, but I had some spare time...don't judge me!_).

I was having a really hard time figuring out the scheme of it until I tried it with `{mode:'cpu'}` where suddenly everything acted exactly as I had expected all along (which it previously had not been). So I set up a side-by-side test and sure enough it mapped completely different in cpu vs gpu modes (with the gpu mode definitely being the non-expected output).

-------------

Live fiddle of the issue (writes to console, so open dev tools): https://jsfiddle.net/39ztrpvs/

Code:
```javascript
var cpu = new GPU({mode:'cpu'});
var gpu = new GPU({mode:'gpu'});

function testFunc(inp) {
	return inp[this.thread.y][this.thread.x];
}

const cpuKernel = cpu.createKernel(testFunc).setOutput([3, 3]);
const gpuKernel = gpu.createKernel(testFunc).setOutput([3, 3]);

console.log('cpu:', cpuKernel( [[0,1,2], [3,4,5], [6,7,8]] ) );
console.log('gpu:', gpuKernel( [[0,1,2], [3,4,5], [6,7,8]] ) );
```

My Output (on my machine at least: Windows, Chrome):
```javascript
// matches expectation
cpu: [[0, 1, 2],
      [3, 4, 5],
      [6, 7, 8]]

// notice the left column is all messed up
gpu: [[0, 1, 2],
      [0, 4, 5],
      [3, 7, 8]]
```

Specifically (in gpu mode):
Position (0,0) in the input for some reason gets mapped to _both_ (0,0) and (0,1) in the output.
Position (0,1) in the input gets mapped to (0,2) in the output.
Position (0,2) doesn't get mapped to the output at all.

This same pattern of the 1st column doing 0 twice, then continuing normally (then, of course, never getting to the last index) seems to continue regardless of how long I make that y-axis.

fuzzie360:
Interesting! Thanks for the bug report, its very helpful for ensuring compatibility.

The output is correct on my computer. What GPU hardware are you currently using? If you have dual graphics, please list both of them.

BryanGrezeszak:
[...]

PR #329, opened 2018-07-02
fix problem with thread.y and z index being incorrect in some hardware

nickhingston:
Hi, 

Managed to find a couple of hours to look at this issue...

please find a fix for #241 and #251 

Test case for this issue added too - but i don't seem to be able to run the full suite! If I comment out all other tests bar mine its ok...but this kinda defeats the point of a test suite!

I get "Too many active WebGL contexts. Oldest context will be lost." - if one of these has happened, my (and some others) will fail...

New issue?! (it's broken for me without my changes)


test/all.html
```diff
@@ -58,6 +58,7 @@
   <script src="issues/207-same-function-reuse.js"></script>
 	<script src="issues/212-funky-function-support.js"></script>
   <script src="issues/233-kernel-map-float-output.js"></script>
+  <script src="issues/241-CPU-vs-GPU-maps-output-differently.js"></script>
```
robertleeplummerjr: Can you push up your test?

robertleeplummerjr:
Two requests:
* Can you wrap the test in a `(function() { [test here] })();` like the other tests?  When I run the test, `input` gets overwritten and the test sometimes fails.
* Can you use `var` in tests?  We try and have tests pass in older versions of browsers (like modern safari), so we can continue to test in the duration we need to?

Fantastic work!

nickhingston:
sure, you want me to fix 259-atan2.js too, which is where i templated it from?!

i guess there are others 263- etc...

seems there are other es6 features in some of these tests too...  e.g. () => {

Looks like you need a separate task to sort out the test framework a bit, but will fix this file as its new.

robertleeplummerjr:
@nickhingston, fantastic job!  I look forward to more!

PR #333, opened 2018-07-03
allow for 2^24 addresses (was 2^23)

### LINKED ISSUE #314: Cant address input array with index >= 2^23
I have a problem with addressing an input array at 8388608 or higher.
the resultant value is incorrect - i think it is returning a value a few indexes left or right of the required element (as image is kinda recognisably corrupt after this point).  I guess some integers with Float32 overflow issue.

Using cpu mode fixes the problem.

I tried breaking up the array into a 2d array, and this improved things, but have a similar issue when the combined index >= 2^24 - so gets twice as far before image corrupting.

so...i tried a 3d array...same result as 2d.

Finally, I have it working but had to send the 3d array as separate parameters in the kernel function.
e.g.

myKernelFunction(threeDArray[0], threeDArray[1], threeDArray[2], ...)

which makes my code more complex than it needs to be, esp as in the kernel i cant seem to assign a variable to the correct array part.  I guess all variables created by the shader compiler are Float32?

nickhingston:
Partially fixes #314, but we should be able to address with int32 i.e. 2^32 but think this may be a big change...but kinda important if you have a file > 16MBytes to operate on...a use case I certainly have.

Any idea why we are using the 'thread' as a vec3 vector, why not uvec3 and do all addressing with integers?


nickhingston:
sorry - conflicts just dates in the built code...

nickhingston:
i guess there could be, cos we are doing a load of int -> float conversions and back again!

should certainly help with code simplicity, and rounding bugs we are seeing!

looking into it...

robertleeplummerjr:
Lets handle the int change over in another issue, and go ahead and try and get these other solved asap.

PR #339, opened 2018-07-05
314 large input array addressing

nickhingston:
use ints for this.thread.x/y/z!

big change...but worth it to fix my #314 problem completely.

Dunno of the performance impact...might be worth a quick investigation...


robertleeplummerjr:
updated https://github.com/gpujs/gpu.js/tree/314-large-input-array-addressing with your changes.  Before the changes, if I regularly run the suite I get:
```
301 tests completed in 11243 milliseconds, with 0 failed, 0 skipped, and 0 todo.
1115 assertions of 1115 passed, 0 failed.
```

After the changes, if I regularly run the suite I get:
```
```

robertleeplummerjr:
My first hunch is `debug_log`

robertleeplummerjr:
If I get you commit access, you think you can branch of `develop` and we can work on these issues together?

robertleeplummerjr:
I should clarify: _future_ branches would go off of develop, that'd keep conflicts down.

This should continue on `314-large-input-array-addressing`

nickhingston:
i think the 'compile' side of things *will* be slower - because its more complex having to deal with the different states!

But the reason for the large difference in performance just running the test cases between the 2 is this line in 

test/issues/314-large-input-array-addressing.js

`var DATA_MAX = 8388800; `

changed to

`var DATA_MAX = 8388800*8;` !

notice the *8 - which really is a very large array!

yeah, be happy to added to the team....i guess we should have a bit of a chat soon 👍

Issue #300, opened 2018-05-29
nested array index does not work in kernel function


sroboubi:
var gpu = new GPU();

        var broken = this.gpu.createKernel(function(input, lookup)
        {
            return lookup[input[this.thread.x]];
        }).setOutput([1]);

        var working = this.gpu.createKernel(function(input, lookup)
        {
            var idx = input[this.thread.x];
            return lookup[idx];
        }).setOutput([1]);

        console.log(broken([2], [7, 13, 19, 23]));
        console.log(working([2], [7, 13, 19, 23]));

I would expect both kernels to return 19, but the first one (broken) returns 2. Is this a bug?

nickhingston:
this is not a bug.

you can't pass a function to the kernel as a parameter - this is not a 'nested' function.

The function needs to be compiled to shader language at 'compile time', you are passing this function in at runtime.

See createKernelMap documentation - i think this is what you are after...

robertleeplummerjr:
The differences in syntax of

`return lookup[input[this.thread.x]];`

and

```
var idx = input[this.thread.x];
return lookup[idx];
```

shouldn't cause there to be two different results.  I still see this is as a bug.

nickhingston:
i think i added this comment to the wrong issue somehow!

robertleeplummerjr:
In either case, it is a pleasure to work with another human @nickhingston ;)

fuzzie360:
Closing the loop on this seven years later: it was a real bug, and it is now fixed in `develop` (02b346c).

**What was happening.** `lookup[input[this.thread.x]]` compiles to a texture read nested inside another texture read's argument list — both through the same generated helper that takes a `sampler2D` parameter:

```glsl
getMemoryOptimized32(user_lookup, ..., int(getMemoryOptimized32(user_input, ..., threadId.x)))
```

[...]

PR #343, opened 2018-07-10
314 large input array addressing

robertleeplummerjr:


src/backend/web-gl/function-node.js
```diff
@@ -11,6 +11,10 @@ const constantsPrefix = 'this.constants.';
 const DECODE32_ENCODE32 = /decode32\(\s+encode32\(/g;
 const ENCODE32_DECODE32 = /encode32\(\s+decode32\(/g;
 
+// these debugs were hugely usefull...
+// TODO: optimise out - webpack/babel options? maybe some generic logging support in core/utils?
+// const debug_log = console.log
+const debug_log = () => {};
```
robertleeplummerjr: I think it really depends on what is being debugged.  We generally remove debugging as much of this can be done using modern debugging tools.
At the very least, it should follow casing already defined `debugLog`.  If we used it as a method, we could check if debug is enabled inside it, and only do something if debug as enabled.  Example:

```js
debugLog(msg) {
  if (!this.debug) return;
  // do stuff here
}
```

Then later:
```js
this.debug('something');
```

However, this really is outside the scope of this issue, and if it were introduced, it'd be best to open an issue, and address it there.  In the mean time, removing it here.

src/backend/web-gl/function-node.js
```diff
@@ -177,14 +181,14 @@ module.exports = class WebGLFunctionNode extends FunctionNodeBase {
 	 */
 	astReturnStatement(ast, retArr) {
 		if (this.isRootKernel) {
-			retArr.push('kernelResult = ');
+			retArr.push('kernelResult = float(');
```
robertleeplummerjr: Couldn't we move the casting into the js side, rather than it being casted in glsl?

dist/backend/web-gl/function-node.js
```diff
@@ -868,14 +890,20 @@ module.exports = function (_FunctionNodeBase) {
...
 						// Get from array
 						this.astGeneric(mNode.object, retArr);
 						retArr.push('[int(');
 						this.astGeneric(mNode.property, retArr);
 						retArr.push(')]');
 					} else {
+						var multiMemberExpression = this.isState('multi-member-expression');
+						if (multiMemberExpression) {
+							this.popState('multi-member-expression');
+							this.pushState('not-in-get-call-parameters');
```
robertleeplummerjr: You are pushing a state that is never checked for, so is this actually needed?

src/backend/web-gl/function-node.js
```diff
@@ -848,93 +876,143 @@ module.exports = class WebGLFunctionNode extends FunctionNodeBase {
 							retArr.push('[');
 							retArr.push(mNode.property.raw);
 							retArr.push(']');
+							multiMemberExpression && this.popState('not-in-get-call-parameters');
```
robertleeplummerjr: Can you write this less cryptically?

src/backend/web-gl/function-node.js
```diff
@@ -848,93 +876,143 @@ module.exports = class WebGLFunctionNode extends FunctionNodeBase {
...
 							retArr.push('Dim[1],');
 							this.astGeneric(mNode.object, retArr);
 							retArr.push('Dim[2]');
 							retArr.push('), ');
+							if (multiMemberExpression) {
+								this.popState('not-in-get-call-parameters');
+							}
+							this.pushState('in-get-call-parameters');
 							this.astGeneric(mNode.property, retArr);
+							!multiMemberExpression && this.popState('in-get-call-parameters');
```
robertleeplummerjr: Can you write this less cryptically?

src/backend/web-gl/function-node.js
```diff
@@ -848,93 +876,143 @@ module.exports = class WebGLFunctionNode extends FunctionNodeBase {
...
```
robertleeplummerjr: Can you write this less cryptically?

src/backend/web-gl/kernel-string.js
```diff
@@ -56,8 +56,10 @@ module.exports = function(gpuKernel, name) {
 		    this.uniform1fCache = {};
 		    this.uniform1iCache = {};
 		    this.uniform2fCache = {};
-		    this.uniform2fvCache = {};
-		    this.uniform3fvCache = {};
+        this.uniform2fvCache = {};
```
robertleeplummerjr: spacing is off

src/backend/web-gl/kernel-string.js
```diff
@@ -80,8 +82,10 @@ module.exports = function(gpuKernel, name) {
 		  ${ removeFnNoise(gpuKernel.setUniform1f.toString()) }
 		  ${ removeFnNoise(gpuKernel.setUniform1i.toString()) }
 		  ${ removeFnNoise(gpuKernel.setUniform2f.toString()) }
-		  ${ removeFnNoise(gpuKernel.setUniform2fv.toString()) }
-		  ${ removeFnNoise(gpuKernel.setUniform3fv.toString()) } 
+      ${ removeFnNoise(gpuKernel.setUniform2fv.toString()) }
```
robertleeplummerjr: spacing is off

src/backend/web-gl/kernel.js
```diff
@@ -947,7 +980,7 @@ module.exports = class WebGLKernel extends KernelBase {
 	_getGetTextureIndexString() {
 		return (
 			this.floatTextures ?
-			'  index = float(int(index)/4);\n' :
+			'  index = index / 4;\n' :
```
robertleeplummerjr: very nice revision!

src/backend/web-gl/shader-frag.js
```diff
@@ -38,8 +38,8 @@ highp float integerMod(highp float x, highp float y) {
   return res * (res > floor(y) - 1.0 ? 0.0 : 1.0);
 }
 
-highp int integerMod(highp int x, highp int y) {
-  return int(integerMod(float(x), float(y)));
+highp int integerMod(highp int a, highp int n) {
```
robertleeplummerjr: Can we keep `x` and `y` here?

src/backend/web-gl2/kernel-string.js
```diff
@@ -56,8 +56,10 @@ module.exports = function(gpuKernel, name) {
 		    this.uniform1fCache = {};
 		    this.uniform1iCache = {};
 		    this.uniform2fCache = {};
-		    this.uniform2fvCache = {};
-		    this.uniform3fvCache = {};
+        this.uniform2fvCache = {};
```
robertleeplummerjr: spacing is off

src/backend/web-gl2/kernel-string.js
```diff
@@ -80,8 +82,10 @@ module.exports = function(gpuKernel, name) {
 		  ${ removeFnNoise(gpuKernel.setUniform1f.toString()) }
 		  ${ removeFnNoise(gpuKernel.setUniform1i.toString()) }
 		  ${ removeFnNoise(gpuKernel.setUniform2f.toString()) }
-		  ${ removeFnNoise(gpuKernel.setUniform2fv.toString()) }
-		  ${ removeFnNoise(gpuKernel.setUniform3fv.toString()) } 
+      ${ removeFnNoise(gpuKernel.setUniform2fv.toString()) }
```
robertleeplummerjr: spacing is off

src/backend/web-gl2/kernel.js
```diff
@@ -608,7 +608,7 @@ module.exports = class WebGL2Kernel extends WebGLKernel {
 				}
 
 				if (i < channels.length - 1) {
-					result.push('  index += 1.0');
+					result.push('  index += int(1)');
```
robertleeplummerjr: You are casting an int, to an int?

test/issues/314-large-input-array-addressing.js
```diff
@@ -1,8 +1,9 @@
 (function() {
-	// max size of ok addressing = 8388608, 8388609 is shifted by 1 so index seems to be 8388610
-	// after this fix max addressing is 2^24 which is the max a float32 can handle
-	// question is, why cant we use int32? so 2^32 addressing...
-	var DATA_MAX = 8388800;
+	// max size of ok addressing was 8388608, 8388609 is shifted by 1 so index seems to be 8388610
+	// after this fix max addressing is 2^31 which is the max a int32 can handle
+	// run out of heap before being able to create a butter that big!
+	// wanted to use uints but caused more problems than it solved
+	var DATA_MAX = 8388800*8;
```
robertleeplummerjr: amazing!

src/backend/web-gl/function-node.js
nickhingston: it is 'removed' here. its just an empty function...you need to uncomment line 16 for debug to be enabled!

I don't know any modern debugger features that can give this kind of info without lots of faffing about - you just don't get the 'history' needed.  It was hugely useful again for fixing #300, just switched it on and the problem easy to find...

so I'm pretty reticent about removing it as I know I will need to stick it back in again!

dist/backend/web-gl/function-node.js
nickhingston: yes it is! We need the state that we are no longer currently in the 'in-get-parameters' state ( I was thinking of calling it the force-int state which might help with your understanding - but probably too generic for the actual lib).

any call to this.astGeneric will check the state and nested gets will have issues without this...

src/backend/web-gl/function-node.js
nickhingston: you mean:
if (multiMemberExpression) {
  this.popState('not-in-get-call-parameters');
}

pretty standard in JS! kinda powerful...but happy to rewrite.  You got a coding standard?

I think this is an eslint option you can turn on/off - i think AirB&B eslint standard has it off which has annoyed be before!

src/backend/web-gl/shader-frag.js
nickhingston: sure

src/backend/web-gl2/kernel-string.js
nickhingston: hmmm, tabs v spacing?!

src/backend/web-gl/function-node.js
robertleeplummerjr:
```js
if (multiMemberExpression) {
  this.popState('not-in-get-call-parameters');
}
```

Yes, this would be more straightforward to the next human and is the existing standard in the library thus far.

src/backend/web-gl2/kernel.js
nickhingston: yeah, unnecessary! will fix

src/backend/web-gl/function-node.js
nickhingston: yes, it seems that this is unnecessary - i suppose i wanted to catch the possibility of `return this.thread.z` which is a pretty unlikely and pointless scenario thinking about it.  All test cases pass without it!

if it comes a problem in future I guess we need a 'force-float' state and add a cast if necessary

nickhingston:
ok, think this is it...if you happy with these?

robertleeplummerjr:
Fantastic work!

PR #351, opened 2018-07-25
fix integer division issues on some cards

### LINKED ISSUE #349: Test case 'failure' - FP accuracy issue
screen shot 2018-07-23 at 14 54 59

is encode32 & decode32 working ok?

nickhingston:
Fixes #349 - well, works around my 'buggy' apple hardware!

also removed any references to highp in generated GLSL as the default precision is already specified as highp

put a note in readme on what it's for, but performance hit is pretty much imperceptible on my HW!


robertleeplummerjr:
I think we should actually not force the user to run this, but rather have this as perceptible startup that can be detected and automatically handled.

nickhingston:
the user has to decide wether to use this.

we cant auto detect at ‘compile' time, because you need a runtime check for the HW (if we are ever going to properly support the gpu core only setup - which would be nice and probably my next issue to look at)

robertleeplummerjr:
>the user has to decide wether to use this.

Why cannot we detect this here, like we do all the others?: https://github.com/gpujs/gpu.js/blob/develop/src/core/utils-core.js#L254

nickhingston:
because you don't know if you have the bug until you run a kernel and try a div by 3.

robertleeplummerjr:
We can do that, just like here: https://github.com/gpujs/gpu.js/blob/develop/src/core/utils.js#L317 and used here https://github.com/gpujs/gpu.js/blob/eb24dda5940a743b70fc326ba50e36de251ce550/src/backend/web-gl/kernel.js#L86

nickhingston:
OK, not really looked into the core build yet,  i don't really understand the point of it then...my thoughts that it was to get rid of acorn and any ‘compilation’ phase whatsoever, i.e. you’d pass it a precompiled kernel to run…I personally think that should be the goal.

robertleeplummerjr:
If you move the option into utils, as a simple kernel that runs once on initial load of the library that sats a value once, like here: https://github.com/gpujs/gpu.js/blob/develop/src/core/utils.js#L317 and call it like this: ```js
if (_isFloatReadPixelsSupported !== null) {
			return _isFloatReadPixelsSupported
		}
```
and simply reference it here: https://github.com/gpujs/gpu.js/pull/351/files#diff-55e78093e7a49fe7d3c6e2334e6888ccR280 and anywhere else it is needed, and we remove this as an option of the api, it keeps the api clean, and puts the work onto gpu.js so as to figure out if this is needed or not by the environment.

The core build could use some tlc, breaking out acorn, etc. but it is really a different issue than this.

robertleeplummerjr:
This is a design pattern that was set when I arrived at gpu.js, and I don't think we should deviate.

nickhingston:
ok, done & agree core changes should be done under different task.  Though this change will make that task a bit more difficult!

PR #355, opened 2018-08-04
Add initial support for OffcreenCanvas

sandeepmistry:
These changes add support for OffscreenCanvas as discussed in #5.

Some browsers support it via flags: https://caniuse.com/#feat=offscreencanvas

Chrome Canary 70, appears to enable the flags or support  by default.

### DIFF
--- a/src/core/utils-core.js
+++ b/src/core/utils-core.js
@@ -35,21 +35,23 @@ class UtilsCore {
 	 * @memberOf UtilsCore
 	 *
 	 *
-	 * @desc Return TRUE, on a valid DOM canvas object
+	 * @desc Return TRUE, on a valid DOM canvas or OffscreenCanvas object
 	 *
 	 * Note: This does just a VERY simply sanity check. And may give false positives.
 	 *
 	 * @param {CanvasDOMObject} canvasObj - Object to validate
 	 *
-	 * @returns {Boolean} TRUE if the object is a DOM canvas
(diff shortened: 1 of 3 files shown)


robertleeplummerjr:
Can you add some test cases perhaps and potentially an option for turning it on or off?  Super awesome feature, cannot wait to get it in!

robertleeplummerjr:
@nickhingston and I were talking about this in private, and he mentioned you already have the option, which is the document itself here: `const canvas = typeof document !== 'undefined' ? document.createElement('canvas') : new OffscreenCanvas(0, 0);`

Pure genius.  Would you be up for adding a quick unit test and possibly documentation?

sandeepmistry:
I've added a test in https://github.com/gpujs/gpu.js/pull/355/commits/aad86a86ab80d3ad050b9f2e8e6a0e2fcf09f80f that only runs if Web Worker + OffscreenCanvas is supported by the browser. It creates a worker and checks the mode. Let me know what you think.

As for the documentation any place in particular you would like to see it? Maybe we can add a snippet of how to use in a Web Worker?

```javascript
var window = {};
importScripts('path/to/gpu.js');

const gpu = new window.GPU();

onmessage = function(e) {
  var input = e.data;
  var output;

  // process input ...
  
  postMessage(output);
};
```

ubernaut:
This is awesome. I knew if I waited long enough someone would do it. thanks @sandeepmistry

robertleeplummerjr:
Merging.  I'll take care of the cleanup.  Awesome work!

PR #356, opened 2018-08-04
Use WebGLRunner if canvas does not support WebGL2

sandeepmistry:
I'm using a old 2011 MacBook Air, with the `#ignore-gpu-blacklist` enabled on Chrome.

`WebGL2RenderingContext` is defined, however the Canvas doesn't support WebGL2 (`canvas.getContext('webgl2')` returns `null`).

These changes check if WebGL2 is supported on the canvas before selecting the `WebGL2Runner`.

### DIFF
--- a/src/core/gpu.js
+++ b/src/core/gpu.js
@@ -261,7 +261,7 @@ class GPU extends GPUCore {
 
 
 	getGPURunner() {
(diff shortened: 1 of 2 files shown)


robertleeplummerjr:
Nice find!

PR #389, opened 2018-10-28
Deep type detection

robertleeplummerjr:


robertleeplummerjr:
@kamil-kielczewski may I have a review?

robertleeplummerjr:
Passing unit tests in Chrome and FF, fyi on a Macbook pro.

README.md
```diff
@@ -360,6 +362,30 @@ const kernel = gpu.createKernel(function(a, b) {
...
 ```
 
+### Adding strongly typed functions
+
+To strongly type a function you may use options.  Options take an optional hash values:
+`returnType`: optional, defaults to float, the value you'd like to return from the function
+`paramTypes`: optional, defaults to float for each param, a hash of param names with values of the return types
+
+Types: that may be used for `returnType` or for each property of `paramTypes`: 
+'Array'
```
kamil-kielczewski: What is the meaning of "Array" type without number o elements? Can we give some example of that type?

dist/backend/web-gl/function-node.js
```diff
@@ -652,17 +683,64 @@ module.exports = function (_FunctionNodeBase) {
...
+					if (init) {
+						if (init.object) {
+							if (init.object.type === 'MemberExpression' && init.object.object) {
+								if (init.object.object.type === 'Identifier') {
+									var _type2 = this.getParamType(init.object.object.name);
+									declarationType = typeLookupMap[_type2];
+									if (!declarationType) {
+										throw new Error('unknown lookup type ' + typeLookupMap);
+									}
+									debugger;
```
kamil-kielczewski: Why we have 'debugger' statements in public code? May be we should remove it? (it exists in lines: 698, 705, 713, 720... and maybe somewhere else...

dist/backend/web-gl/function-node.js
```diff
@@ -889,7 +967,9 @@ module.exports = function (_FunctionNodeBase) {
 		value: function astMemberExpression(mNode, retArr) {
 			debugLog("[in] astMemberExpression " + mNode.object.type);
 			if (mNode.computed) {
-				if (mNode.object.type === 'Identifier' || mNode.object.type === 'MemberExpression' && mNode.object.object.object && mNode.object.object.object.type === 'ThisExpression' && mNode.object.object.property.name === 'constants') {
+				if (mNode.object.type === 'Identifier' || mNode.object.type === 'MemberExpression' &&
+				// mNode.object.object &&
+				mNode.object.object.object && mNode.object.object.object.type === 'ThisExpression' && mNode.object.object.property.name === 'constants') {
```
kamil-kielczewski: mNode.object.object.object.type -> it looks scary :P (however if it works it is no problem for me :P )

src/backend/web-gl/function-node.js
```diff
@@ -642,17 +644,78 @@ module.exports = class WebGLFunctionNode extends FunctionNodeBase {
...
+									init.object.object.object.property.name === 'constants'
+								) {
+									const type = this.getConstantType(init.object.object.property.name);
+									declarationType = typeLookupMap[type];
+									if (!declarationType) {
+										throw new Error(`unknown lookup type ${ typeLookupMap }`);
+									}
+									debugger;
+								} else if (
+									init.object.object.object.object.object.type === 'ThisExpression' &&
```
kamil-kielczewski: init.object.object.object.object.object.type -> this is also very sacry :P

README.md
kamil-kielczewski: I also wann't to say that I don't deeply understand code so I can check on the the surface JS level only

robertleeplummerjr: >What is the meaning of "Array" type without number o elements? Can we give some example of that type?

It is a standard array, passed from javascript to webgl.

dist/backend/web-gl/function-node.js
robertleeplummerjr: >mNode.object.object.object.type -> it looks scary :P (however if it works it is no problem for me :P )

Should be quite a bit easier to read now.  Let me know if that is more acceptable.

robertleeplummerjr:
FYI unit tests passing in FF and Chrome.

dist/backend/web-gl/function-node.js
kamil-kielczewski: I wonder why this object is so nested?

PicoCreator: Noticed i got summoned to this review, might be a little bit too out of touch

But yes I do remember this scary little piece of code, haha. Definitely agree it should be rewritten to be clearer. But im quite sure at that time we were smashing the code together till it works.

```suggestion
if (
	( mNode.object.type === 'Identifier' || mNode.object.type === 'MemberExpression' ) &&
	( mNode.object.object && mNode.object.object.object && 
	  mNode.object.object.object.type === 'ThisExpression' && 
	  mNode.object.object.property.name === 'constants' )
) {
```

Do you think this would be clearer?

PicoCreator: Btw @robertleeplummerjr - seems like I started having more time on github these days (for various other projects), feel free to drag me in for small reviews if you like.... hopefully I still remember something 😂 : you probably know the library better then I do now

robertleeplummerjr: Dist files are built.  Can you apply them to src?  Also welcome back!

robertleeplummerjr:
My intial thought on the proposed code is that it is shorter, but the beforehand the files are more human readable.

robertleeplummerjr:
Also, I am reading this between flights, on my phone, which isn't forgiving for a diff.



Taking off now....

src/backend/web-gl/function-node.js
PicoCreator: Haha, yea. In most of the cases is usually a quick work around AST nesting....

So i might be completely wrong (out of touch), for example `var foo = function foo(a) { ... }`

Because we cant do an AST scan using `function foo(a) {...}` itself, we did a hack to wrap it around into a `var foo`.

We then do an AST scan on that, so if we want to get the function content, it would be.

var -> foo -> assignment -> function -> function body, or something like that.

Hence the `object.object.object.object` because the `var foo = ` part is static which we defined.

PicoCreator:
No worries, have a safe flight. In general I always vote for code readability, as machines will minify it anyway =P

kamil-kielczewski:
I think that code refactoring can be separate task - "TODO" in future (not in this pull request)

robertleeplummerjr:
Going to go ahead and merge, and if the portability of the code becomes an issue in the `astVariableDeclaration` method, we can raise another issue to cleanup.

Ty everyone for your feedback!

PR #428, opened 2019-02-28
fix: Implicit 'any' types

jcaru:
Fix various TypeScript compile errors, all of which relate to implicit 'any' types in index.d.ts


robertleeplummerjr:
Nice job!

PR #458, opened 2019-04-28
Fix return of >=2d arrays from GPU

Drenata:
The return-arrays.js tests are updated to have non-square output sizes and for each output to have a unique value.

I also renamed all variables (which I assume are supposed to be the same thing) to `xResultsMax`.


robertleeplummerjr:
Great find and unit tests

Issue #468, opened 2019-05-17
Crash on installation: error: no matching function for call to ‘v8::FunctionTemplate::GetFunction()’


sbrl:
![A GIF or MEME to give some spice of the internet](url)

## *What* is wrong?
I can't install gpu.js on Ubuntu 18.10.

## *Where* does it happen?
In the terminal, if I type this: `npm install gpu.js`

## *How* do we replicate the issue?
On an Ubuntu 18.10 machine, do `npm install gpu.js`.

## *How* important is this (1-5)?
I'd assume it's a pretty game-breaking issue. 4-5.

## Expected behavior (i.e. solution)
I should be able to install it

## Other Comments
Output:

```
  CXX(target) Release/obj.target/translator_static/angle/src/compiler/translator/ShaderVars.o
  AR(target) Release/obj.target/angle/src/translator_static.a
  COPY Release/translator_static.a
  CXX(target) Release/obj.target/webgl/src/bindings.o
In file included from ../src/webgl.h:10,
                 from ../src/bindings.cc:9:
../../nan/nan.h: In function ‘void Nan::AsyncQueueWorker(Nan::AsyncWorker*)’:
../../nan/nan.h:2298:62: warning: cast between incompatible function types from ‘void (*)(uv_work_t*)’ {aka ‘void (*)(uv_work_s*)’} to ‘uv_after_work_cb’ {aka ‘void (*)(uv_work_s*, int)’} [-Wcast-function-type]
     , reinterpret_cast<uv_after_work_cb>(AsyncExecuteComplete)
                                                              ^
../src/bindings.cc: In function ‘void Init(Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE)’:
../src/bindings.cc:482:35: error: no matching function for call to ‘v8::FunctionTemplate::GetFunction()’
     , webgl_template->GetFunction());
                                   ^
In file included from /home/sbrl/.node-gyp/12.2.0/include/node/node.h:63,
                 from ../src/webgl.h:9,
                 from ../src/bindings.cc:9:
/home/sbrl/.node-gyp/12.2.0/include/node/v8.h:5947:46: note: candidate: ‘v8::MaybeLocal<v8::Function> v8::FunctionTemplate::GetFunction(v8::Local<v8::Context>)’
   V8_WARN_UNUSED_RESULT MaybeLocal<Function> GetFunction(
                                              ^~~~~~~~~~~
[...]

PR #469, opened 2019-05-17
README.md: Add Linux installation notes

sbrl:
This PR adds a small note for Linux users, as there are some non-obvious dependencies which need to be installed in order for the build to succeed - and the error messages are  _singularly unhelpful_.

Note that there _may_ be additional packages missing from the list here, as I've run into some issues and opened #468.


robertleeplummerjr:
Nice work.

PR #506, opened 2019-09-26
Fix calls to FunctionNode.astErrorOutput, it's not a constructor

gnandretta:
Sorry I don't have a repro steps, but after upgrading from v1 I'm getting `TypeError: this.astErrorOutput is not a constructor`.


robertleeplummerjr:
fantastic! Ty for fixing.

PR #518, opened 2019-10-18
add browser field to package.json

### LINKED ISSUE #509: Add browser field to package.json
## *What* is wrong?
When bundling gpu.js for the browser it may throw an error because the 'fs' module of Node.js is required somewhere and is not available in the browser.

## *Where* does it happen?
This is caused by the `bindings` module.

gpu.js does already provide a special build which can used in a browser but unfortunately there is no way for webpack to pick it up yet. What about adding a `browser` field to the package.json file which points to `dist/gpu-browser.js`?

Please let me know if you think this makes sense as well. I would be happy to provide a pull request which adds the `browser` field.

chrisguttandin:
This pull request add a browser field to the package.json which points to the browser bundle.

It fixes #509.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   ],
   "unpkg": "./dist/gpu-browser.min.js",
   "jsdelivr": "./dist/gpu-browser.min.js",
+  "browser": "./dist/gpu-browser.js",
   "directories": {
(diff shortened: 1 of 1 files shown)


robertleeplummerjr:
Nice work!

Issue #521, opened 2019-10-20
strange canvas output if graphical is true on chrome for android


PipOlt:
## *What* is wrong?
I tried to run your yellifish canvas direct draw example on my mobile phone (Samsung Galaxy S7) with latest Android updates. If CPU mode is enabled everything works like expected (slow). If I enable the GPU the output is shuffled. It seems like the texture coordinates are shifted. 
expected (cpu):
![screenshot_1](https://user-images.githubusercontent.com/16139246/67157780-4f3f3400-f331-11e9-97c6-8ed653f0a58a.jpg)
observed (gpu):
![screenshot](https://user-images.githubusercontent.com/16139246/67157806-ae04ad80-f331-11e9-9c75-c1adf6f3e44a.jpg)

I tried another example where an color gradient is drawn directly to the canvas. Same behavior on this specific device.
expected (cpu):
![screenshot_2](https://user-images.githubusercontent.com/16139246/67157982-c4f7cf80-f332-11e9-8500-6b83934524da.jpg)
observed (gpu):
![screenshot_3](https://user-images.githubusercontent.com/16139246/67157993-f2447d80-f332-11e9-9cd7-21a644b133a7.jpg)

## *Where* does it happen?
Chrome for Android on my Samsung Galaxy S7. Didn't try another Android Device. Can try it tomorrow. Linux Desktop works like expected.

## *How* do we replicate the issue?
If you have such a device, you can use the Yellifish Video example. Or you can use the second example below:
```
<!DOCTYPE html>
<meta charset="utf-8">

<body>
<script src="js/gpu.js"></script>
<script>
const gpu = new GPU({mode: "gpu"});
const render = gpu.createKernel(function(x) {
  this.color(this.thread.x/500, this.thread.y/500, x[0]);
})
  .setOutput([500, 500])
  .setGraphical(true)
render([0.4])
const canvas = render.canvas;
document.getElementsByTagName('body')[0].appendChild(canvas);
</script>
```
## *How* important is this (1-5)?
5 (my subjective importance) because I cannot use gpu.js for my project with this behaviour. I want to apply video effects to the device camera. Without this behaviour the performance of gpu.js would be great for this project. It is an awesome library! Thanks for that! :)

## Expected behavior (i.e. solution)
[...]

PR #525, opened 2019-10-29
Second working port to Rollup

robertleeplummerjr:
I merged https://github.com/gpujs/gpu.js/pull/499 too early, thinking that the node unit testing side was correctly implemented.  Because of how drastic the change was, and that it no longer allowed testing in node, I reverted using git checkout HEAD~1 and a force push.  I then reopened this in hopes that the changes will handle node eventually, and to keep a PR open and actionable.


Symbitic:
Tests should work in Node (I will double check). It just imports from the bundled commonjs file (in dist). And Node tests have to use `require` - that's why we import from the bundle instead of src.

For tests, I'm still looking into it. WebGL in browser is notoriously difficult to unit tests. It's even more difficult when there's hundreds of unit tests, like here. Plus, this is testing GPGPU math, not visual output to a canvas.

Some options include:
* Puppeteer, either standalone or with Jest preset and/or environment.
* Electron (mocha is often used to test WebGL this way)
* TestCafe
* Cypress

I'm trying to evaluate which one would be best.

P.S. is it too much to ask that v3 only exports the GPU class? I don't think many people use anything other than that.

Symbitic:
You don't have to use double the src files. The QUnit browser tests already use the bundled version found in dist/ (all.html defines require() to simply redirect to an object based on the name).

Both node and browser tests should work.
To run the node tests:
```
yarn test test/features
yarn test test/internal
yarn test test/issues
```
(just like in the official instructions).

To run the browser tests: `yarn watch`
It starts rollup in watch mode, which will automatically start browsersync and open all.html in the web browser.

Both `watch` and `test` automatically run rollup to produce the dist/ bundle, so requiring the bundle to be built is invisible to the user.

I realize that Puppeteer and Electron use Chrome, but Babylon.js, THREE, A-Frame, and every other major WebGL project I know of have no problem with that; It lets them catch enough errors that the exceptional cases when other browsers differ can be handled manually. Though I admit, this one depends more on mathematical correctness than visual output.

robertleeplummerjr:
I'll resolve this: https://github.com/gpujs/gpu.js/issues/521 then I'll work on helping here, as it is the last high priority item.  Sorry to keep this help up.  I did find an issue with build that I've fixed locally, I'll release today that may have contributed to the build issue I was seeing.

Symbitic:
Thanks! I appreciate your interest.

Did you run the tests for both node and browser?

robertleeplummerjr:
Not yet.

robertleeplummerjr:
Is there any interest in continuing this effort?  Quite a few conflicts.

17097231932:
#809

fuzzie360:
Closing after 6+ years: the tree has diverged far beyond rebasing, and the build-tooling question this explored was settled (for now) by #840 keeping the browserify/gulp chain. The repo now has CI on every PR, so if a Rollup port is ever attempted again it will at least get automatic validation.

PR #545, opened 2019-12-05
Check isArray argument type

franciscofabian:
to avoid error for undefined or null arguments

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -110,7 +110,7 @@ const utils = {
    * @returns {Boolean}  true if is array or Array-like object
    */
   isArray(array) {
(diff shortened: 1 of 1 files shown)


robertleeplummerjr:
I'm curious how an undefined or null would be handled here, because I don't believe those are valid inputs.  Perhaps something more substantial of an error could be thrown indicating an invalid type?

franciscofabian:
I had a null value there and got an error, that’s why I created this PR to fix that error. I’m not sure where that null argument comes from.

robertleeplummerjr:
Perhaps rather than handling the error, it may be best to throw something more descriptive?

PR #583, opened 2020-03-18
Add support for type Uint8ClampedArray in KernelVariable

sarthak-saxena:
```
render = new GPU.GPU({ mode: "gpu" })
render(image.data, 14 * Math.sin(Date.now() / 400))
```
In the above mentioned example since image data is of type Uint8ClampedArray & KernelVariable does not support the same hence gives type error

Ex: https://observablehq.com/@fil/image-to-gpu

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -407,6 +407,7 @@ export type KernelVariable =
   | Uint8Array
   | Uint16Array
   | Uint32Array
+  | Uint8ClampedArray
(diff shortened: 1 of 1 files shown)


robertleeplummerjr:
Released: https://www.npmjs.com/package/gpu.js/v/2.8.5

sarthak-saxena:
> Awesome work!

Thanks :D

PR #595, opened 2020-04-04
Update README.md

GirkovArpa:
Added Caesar Cipher example to demonstrate how string manipulation can take advantage of parallelism.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,25 @@ const multiplyMatrix = gpu.createKernel(function(a, b) {
 
 const c = multiplyMatrix(a, b);
 ```
+```js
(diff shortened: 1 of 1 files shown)


robertleeplummerjr:
Can we move this into an observablehq.com rather than put it directly in the documentation?  I LOVE the example, but the readme is more for focused on finite documentation.

robertleeplummerjr:
I added this, with full credit: https://observablehq.com/@robertleeplummerjr/caesar-cipher-gpu-js-example.  We could then make this PR (or a follow up) the addition of an example.

robertleeplummerjr:
Added here: https://github.com/gpujs/gpu.js/commit/5397186a7ae1e901f504f7682be684cc5572d2c5

PR #631, opened 2020-08-03
Add Skypack to README

FredKSchott:
[Skypack](https://www.skypack.dev) is a popular CDN for importing packages directly in the browser with ESM import syntax. For a lot of users, this syntax is more explicit and more familiar than the older, implicit script tag.

I added a full import snippet here for easier copy-pasting, but also happy to replace with just the URL.

You can also run this quickly in your browser devtools via: `const pkg = await import('https://cdn.skypack.dev/gpu.js');`


fuzzie360:
Closing as stale: Skypack has been effectively frozen for years, so recommending it to new users would be a disservice. The README now documents unpkg/jsDelivr; an esm.sh line would be a welcome fresh PR if ESM-CDN usage is worth documenting.

Issue #589, opened 2020-03-26
Angular build fails (some extra commas in arrays)


Martin-Antonov:
![A GIF or MEME to give some spice of the internet](https://img.devrant.com/devrant/rant/r_812234_oQpnP.jpg)

## *What* is wrong?
I have successfully and very easily integrated this fantastic library in my angular project. However, running `ng build --prod` throws errors "Unexpected token: punc (()"
This is caused when adding a comma after the last element of an array. 
For example, on line 18696 in 'gpu-browser.js', there is extra comma at the end
`data${i + 1}[2] = subKernelResult_${subKernel.name}[2],` <----
There are around 50 or so similar places throughout the code which I fixed by hand to make the project run. 

## *Where* does it happen?
Locally on, on my windows machine. 

## *How* do we replicate the issue?
1. Install angular cli with `npm install -g @angular/cli`
2. Create a new project `ng new test`, when prompted choose whatever css preprocessor you like
3. Go into your new project folder and type 'npm install --save gpu.js'
4. Go to angular.json and add the following to the scripts array: `"./node_modules/gpu.js/dist/gpu-browser.js"`
5. Go to tsconfig.json and add the following to the typings array: `"node_modules/gpu.js/src/index.d.ts"`
6. Go to app.component.ts and import gpu.js via `import {GPU} from 'gpu.js'`
7. Create a simple example using the library
8. Run `ng build --prod --source-map=true --build-optimizer=false`, where the last 2 options will show you the exact lines where the error occurs. 
If you have any problems replicating this, please PM me. 

## *How* important is this (1-5)?
I would say 2. For now, I will copy the modified files in the project, but would love to be able to cleanly npm install and run production for my jenkins builds.

## Expected behavior (i.e. solution)
Be able to easily install and run gpu.js in an Angular app. 

## Other Comments
Thank you for your time and effort in building this awesome library!

robertleeplummerjr:
Are you sure this wouldn't be an issue in angular?

Martin-Antonov:
@robertleeplummerjr , thank you for your message.
I believe angular simply uses Webpack to build the project, so in theory it should be reproducible with Webpack as well. I have not seen this convention (comma after the last element in an array) in other packages and I have not had such problems with other libraries. 

To sum up, it likely isn't, but I am not such an expert to say so.

evanakm:
[...]

PR #665, opened 2021-01-19
Remove trailing commas

evanakm:
Fix for https://github.com/gpujs/gpu.js/issues/589

For an Angular 7.1 project, the prod build doesn't work, and it is caused by terser at the backend. Specifically the issue described here:

terser/terser#412

There are several examples in GPU.js of result.push('foo','bar','quux',) (note the trailing comma). These need to be removed. I have manually removed them in my local copy of the package (in dist/gpu-browser.js and dist/gpu-browser-core.js) and it builds. These seem to come from several of the source files in the back end. It should be an easy fix.

### DIFF
--- a/dist/gpu-browser-core.js
+++ b/dist/gpu-browser-core.js
@@ -11229,7 +11229,7 @@ float integerCorrectionModulo(float number, float divisor) {
     result.push(
       '  threadId = indexTo3D(index, uOutputDim)',
       '  kernel()',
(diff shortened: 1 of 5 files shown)


robertleeplummerjr (approved):
I like it!

robertleeplummerjr:
Released in https://github.com/gpujs/gpu.js/releases/tag/2.11.2

PR #683, opened 2021-03-23
Fix atan2 implementation to behave like Javascript Math.atan2

### LINKED ISSUE #647: atan2 is implemented as atan(y/x) in gpu backend
slow down egghead

## *What* is wrong?
atan2 should return values from -pi to pi, but instead returns values from -pi/2 to pi/2 in gpu mode. It also returns undefined results for x=0.

## *Where* does it happen?
In Chrome browser.

## *How* do we replicate the issue?
I'm attaching a repro case - correct behavior can be seen in mode:cpu, incorrect behavior in mode:gpu.

[atan2-bug.zip](https://github.com/gpujs/gpu.js/files/5571826/atan2-bug.zip)

## *How* important is this (1-5)?

3 - I can work around, but also should be an easy fix.

## Expected behavior (i.e. solution)
The backend GLSL implementations of atan2 use the single-argument version of atan(y_over_x), instead of the two-argument version atan(y, x).

ted-piotrowski:
- Previous implementation returned values in -PI/2 to PI/2 range even though Javascript returns in -PI to PI range
- Add unit tests covering edge cases

Fixes #647 

Previously the following cases failed:

```
Math.atan2(1, 0);
Math.atan2(-1,0);
Math.atan(0,-1);
```

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2) `Math.atan2(y, x)` returns a value between -PI and PI. According to [OpenGL](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/atan.xhtml) `atan(y / x)` returns a value between -PI/2 and PI/2 while `atan(y, x)` returns a value between -PI and PI. 

The main fix was to change the `atan` signature to take `(y, x)` instead of `(y / x)` and then to handle the cases `Math.atan2(-1, 0)`, `Math.atan2(1, 0)`, `Math.atan(0, 0)` because in OpenGL `atan` is undefined when `x = 0` but defined in Javascript.

Note: `atan` arguments are passed `y` first and then `x`. This caused me some confusion at first.

### DIFF
--- a/src/backend/web-gl/fragment-shader.js
+++ b/src/backend/web-gl/fragment-shader.js
@@ -24,8 +24,12 @@ float asinh(float x) {
 }
 
 float atan2(float v1, float v2) {
-  if (v1 == 0.0 || v2 == 0.0) return 0.0;
-  return atan(v1 / v2);
+  if (v2 == 0.0) {
+    if (v1 == 0.0) return 0.0;
+    if (v1 > 0.0) return 1.5707963267948966;
+    if (v1 < 0.0) return -1.5707963267948966;
+  }
+  return atan(v1, v2);
 }
 
 float atanh(float x) {
--- a/src/backend/web-gl2/fragment-shader.js
+++ b/src/backend/web-gl2/fragment-shader.js
@@ -14,8 +14,12 @@ __CONSTANTS__;
 in vec2 vTexCoord;
 
 float atan2(float v1, float v2) {
-  if (v1 == 0.0 || v2 == 0.0) return 0.0;
-  return atan(v1 / v2);
+  if (v2 == 0.0) {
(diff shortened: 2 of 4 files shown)


ted-piotrowski:
@robertleeplummerjr I noticed you've been active on this project again. What are your thoughts on merging this fix?

sojs-coder:
@robertleeplummerjr @ted-piotrowski any update on merging this fix? [this issue](https://github.com/gpujs/gpu.js/issues/828) I believe relies on it.

fuzzie360:
Merged — thank you for the correct fix and the test, and apologies it sat for five years. Released in [2.18.3](https://github.com/gpujs/gpu.js/releases/tag/2.18.3).

PR #723, opened 2021-10-27
Fix divWithIntCheck for floats larger than max int

rotu:
The integer accuracy fix could previously return a wrong result when its operands were bigger than max int.


fuzzie360:
Closing as superseded: the divWithIntCheck rewrite in [2.19.0](https://github.com/gpujs/gpu.js/releases/tag/2.19.0) (for #742) removed the `int()` casts entirely — the divisibility check now rounds the quotient and verifies by exact float multiplication, so the out-of-range undefined behavior this PR targeted no longer exists. (As a side note, `int(x) == x` would not compile on GLSL ES 1.00 — no implicit int→float comparison — so this needed a different shape anyway.) Thanks for flagging the UB.

Issue #727, opened 2021-11-24
Fragment shader functions not in working order


m0ose:
## *What* is wrong?
I have a problem with a shader. The functions in the shader are not in the correct order. I am trying to add roughly 20 functions, and they are breaking the compiled fragment shader. The problematic shader has some functions before functions it depends on, which apparently does not work in webGL. I rearranged the shader functions by hand and it works fine.  

## *Where* does it happen?
A website running GPU.js on my desktop Mac mini (specifically this website https://m0ose.github.io/jsSketches/gpgpuTest/index2.html)

## *How* do we replicate the issue?
I put the example webpage here https://m0ose.github.io/jsSketches/gpgpuTest/index2.html , and the source code for it is here https://github.com/m0ose/jsSketches/tree/master/gpgpuTest .
The first time it tries to run the kernel it uses the compiled shader, and throws an error. Then, it sets the shader to the hand edited one(with the functions in a different order) and it works. 

## *How* important is this (1-5)?
𝜋

## Expected behavior (i.e. solution)
I am not sure what is causing this bug. I was not able to extract a smaller example, which might show the problem better. Could it be caused by the function names?

## Other Comments
This is an Amazing library.

PR #730, opened 2021-12-01
Bug with complicated function dependencies

m0ose:
There was a bug when someone added a bunch of interdependent functions. I think it had to do with the order that they were declared in the shader. 
Here is a fix and a test for #727.

### DIFF
--- a/src/backend/function-builder.js
+++ b/src/backend/function-builder.js
@@ -292,9 +292,6 @@ class FunctionBuilder {
       if (functionIndex === -1) {
         retList.push(functionName);
         functionNode.toString(); //ensure JS trace is done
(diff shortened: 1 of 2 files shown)


fuzzie360:
This looks like a real fix for #727 and the 16-function dependency test is appreciated. One concern before we merge: the reworked traceFunctionCalls recurses into callees unconditionally, so a cyclic call graph (a→b→a) that previously terminated would now recurse infinitely — hanging instead of erroring. Cycles are invalid for GLSL anyway, but a guard (e.g. a currently-tracing set that throws a descriptive error on re-entry) would make this safe. Would you be up for adding that? If not, say the word and we will take it over the finish line — CI now validates every PR automatically.

PR #731, opened 2021-12-02
Bug fix: assign correct argument bit ratio index

### LINKED ISSUE #600: Uncaught Error: argument bit ratio not found
## *What* is wrong?
Adding certain custom functions causes the error: `Uncaught Error: argument bit ratio not found`

## *Where* does it happen?
Not system specific

## *How* do we replicate the issue?
```html
<script>
const gpu = new GPU()

gpu.addFunction(function f(x, y) {
    return y[0]
})

const test = gpu.createKernel(function (a) {
    f(1, a)
}).setOutput([4, 4]).setGraphical(true)

test([0])
</script>
```

Setting `setGraphical(true)` causes the *error* for some reason

## *How* important is this (1-5)?
3 or 4 maybe?

## Expected behavior (i.e. solution)
No error

## Other Comments
In `FunctionBuilder.assignArgumentBitRatio`, replacing `calleeNode.argumentBitRatios[i]` with `calleeNode.argumentBitRatios[argumentIndex]` should fix the problem.

Also, why does having a blank `return` statement cause an error
```js
const test = gpu.createKernel(function () {
    // stuff
    return
}).setOutput([4, 4]).setGraphical(true)
test() // Uncaught Error: Unexpected return statement on ...
```

when there is no error with having no `return` statements

```js
const test = gpu.createKernel(function () {
    // stuff
}).setOutput([4, 4]).setGraphical(true)
test()
```

exebetche:
In "assignArgumentBitRatio" the index for "argumentBitRatios" should be the one of the function, not the kernel. Otherwise, you get a "bit ratio not found" error if you pass arguments to a function with a different order than they're passed to the kernel.
Fix #600

### DIFF
--- a/src/backend/function-builder.js
+++ b/src/backend/function-builder.js
@@ -576,7 +576,7 @@ class FunctionBuilder {
       }
       return calleeBitRatio;
     }
-    calleeNode.argumentBitRatios[i] = bitRatio;
+    calleeNode.argumentBitRatios[argumentIndex] = bitRatio;
     return bitRatio;
   }
 


fuzzie360:
Closing as superseded by #846, which fixes both the read and write sides of this bug and adds tests — but this PR found it first, three and a half years earlier. Thank you; noted in the 2.19.1 release.

PR #809, opened 2023-02-01
Modernization

tq-xyy:
I don't split using branch. Because I didn't notice when I started.

I made the following changes:
* use pnpm to replace with npm
* use prettier as the code formatters but I didn't change code style.
* use es6 module syntax and rollup bundler.
* replace gulp with some modern tools and scripts

The commits are mess, so you can see this pull requests as a issue with its experiment.

You can edit by yourself, even close this pull request . It is up to you. This project has not been updated for a long time, so I want to make it again.


fuzzie360:
Closing per your own framing of this as an experiment we were free to close. The dependency/build modernization goal was reached along a different path in #840 (browserify/gulp retained, released as 2.17.0) — a 300+ file rework against a 3.5-year-old base is no longer mergeable in practice. If ESM/rollup migration becomes a goal, a fresh PR against current develop would be the way. Thanks for the exploration.

PR #824, opened 2023-08-18
Fix cropped rendering

christianh104:
Resolves cropped rendering when WebGL drawing buffer is smaller than canvas due to hardware constraints by scaling viewport to drawing buffer rather than canvas size. See:
https://stackoverflow.com/questions/29710696/webgl-drawing-buffer-size-does-not-equal-canvas-size https://registry.khronos.org/webgl/specs/latest/1.0/#2.2

### DIFF
--- a/src/backend/gl/kernel.js
+++ b/src/backend/gl/kernel.js
@@ -929,9 +929,9 @@ class GLKernel extends Kernel {
       this.updateMaxTexSize();
       this.framebuffer.width = this.texSize[0];
       this.framebuffer.height = this.texSize[1];
-      gl.viewport(0, 0, this.maxTexSize[0], this.maxTexSize[1]);
       this.canvas.width = this.maxTexSize[0];
       this.canvas.height = this.maxTexSize[1];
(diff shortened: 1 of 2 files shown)


fuzzie360:
Thanks for this and for the spec references — we merged it today, and I have to report it didn't survive contact with the test suite: clamping the viewport to `gl.drawingBufferWidth/Height` breaks rendering to framebuffer-attached textures, which don't go through the drawing buffer at all. On headless-gl (where the drawing buffer is fixed at context-creation size) 500+ tests failed, and the same logic would affect any FBO pass whose texture exceeds the initial buffer. It's reverted in [2.19.1](https://github.com/gpujs/gpu.js/releases/tag/2.19.1).

The underlying clamped-canvas bug you describe is real — but the fix needs to distinguish the canvas-output path (where clamping to the drawing buffer is right) from FBO passes (where the viewport must match the texture size). If you're still around and want to rework it along those lines, we'd gladly take it; CI now runs on every PR so it'll get validated automatically.

PR #843, opened 2025-05-20
fix: Update GPU instantiation to use GPU.GPU() in example files

diego-betto:
Fixes "Uncaught TypeError: GPU is not a constructor" error in HTML examples.

### DIFF
--- a/examples/fluid.html
+++ b/examples/fluid.html
@@ -27,7 +27,7 @@
       this.keys = {};
 
       this.canvas = document.createElement('canvas');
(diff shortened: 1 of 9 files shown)


fuzzie360:
Closing as superseded: the root cause (`GPU is not a constructor` in WebGPU-capable browsers, #844) was fixed in the browser bundle itself in 2.17.1, so `new GPU()` in the examples is now the correct working form — switching them to the `new GPU.GPU()` workaround would point users at the old workaround instead. Thanks for the report that helped pin this down.

PR #846, opened 2025-06-05
fix: correct argument index usage in assignArgumentBitRatio

ChenZhuoSteve:
fix: correct argument index usage in assignArgumentBitRatio

Previously, assignArgumentBitRatio incorrectly used the caller's argument index (i) to update the callee function's argumentBitRatios array. This could cause incorrect bit ratio assignment when the caller and callee functions have different argument orders. This commit fixes the bug by using the given callee argument index (argumentIndex) instead.
Added test code.


ChenZhuoSteve:
this is only a issues if the argumentTypes parameter is declared

